PR #1, opened 2020-12-22
Feature/density chart

tdesfont:


resource/dku-helpers.js
```diff
@@ -0,0 +1,93 @@
+/*
+Helper function to query webapp backend with a default implementation for error handling
+v 1.0.1
+*/
+
+function checkMandatoryParameters(param, webAppConfig) {
+    if (param.mandatory) {
+        var val = webAppConfig[param.name];
```
HenriChabert: As said in the call, use `const` or `let` rather that `var`

js/leftBarParamsModule.js
```diff
@@ -0,0 +1,13 @@
...
+app.controller('MyCustomFormController', ['$scope', function($scope) {
+    console.log("Calling the controller");
+    $scope.config.intensity = 50;
+    $scope.config.radius = 50;
+    $scope.config.color = {
+        name: 'blue'
+    };
+    $scope.config.maptile = {
+        name: 'mapbox/v11'
+    };
```
HenriChabert: I would place the default values into constants (for example: `const DEFAULT_INTENSITY = 50`)

python-lib/dku_filtering/filtering.py
```diff
@@ -0,0 +1,102 @@
+import pandas as pd
+from functools import reduce
+
+
+def numerical_filter(df, filter):
```
HenriChabert: name.
```suggestion
def numerical_filter(df, filter_):
```
If possible, it is better not to use builtin names in functions. The general workaround is to use var name `filter_`, that is the main role of placing an underscore at the end of a var

python-lib/dku_filtering/filtering.py
```diff
@@ -0,0 +1,102 @@
...
+    if filter["maxValue"]:
+        conditions += [df[filter['column']] <= filter['maxValue']]
+    return conditions
+
+
+def alphanum_filter(df, filter):
+    conditions = []
+    excluded_values = []
+    for k, v in filter['excludedValues'].items():
+        if k != '___dku_no_value___':
```
HenriChabert: It would be better to use a constant DKU_NO_VALUE for that string.

python-lib/dku_filtering/filtering.py
```diff
@@ -0,0 +1,102 @@
...
+            if v:
+                excluded_values += [k]
+        else:
+            if v:
+                conditions += [~df[filter['column']].isnull()]
```
HenriChabert: Is there a reason you are using a dict for `excludedValues` while you are only using it if the value is True ? Couldn't you just use a list with excluded values?

python-lib/dku_filtering/filtering.py
```diff
@@ -0,0 +1,102 @@
...
+        conditions += [df[filter['column']] >= filter['minValue']]
+    if filter["maxValue"]:
+        conditions += [df[filter['column']] <= filter['maxValue']]
+    return conditions
+
+
+def alphanum_filter(df, filter):
+    conditions = []
+    excluded_values = []
+    for k, v in filter['excludedValues'].items():
```
HenriChabert: nitpicking: If using the var only in a Python script, prefer snake_case to remain consistent.

python-lib/dku_filtering/filtering.py
```diff
@@ -0,0 +1,102 @@
...
+            raise Exception("Unknown date filter.")
+
+    return conditions
+
+
+def apply_filter_conditions(df, conditions):
+    """
+    return a function to apply filtering conditions on df
+    """
+    if len(conditions) == 0:
```
HenriChabert: nitpick: More pythonic to use `if not conditions`

python-lib/dku_filtering/filtering.py
```diff
@@ -0,0 +1,102 @@
...
+        return df
+    elif len(conditions) == 1:
+        return df[conditions[0]]
+    else:
+        return df[reduce(lambda c1, c2: c1 & c2, conditions[1:], conditions[0])]
```
HenriChabert: Nice way to do it!

resource/dku-helpers.js
```diff
@@ -0,0 +1,93 @@
...
+        if (val == undefined || val == "") {
```
HenriChabert:
```suggestion
        if (val === null || val === "") {
```
null returns true for both undefined and null, which is what you want unless you want null to be considered as ok.
You should also use `===` syntax that will run a deep check (verify type and value). For example, 1 == '1' returns true while 1 === '1' returns false. Using `===` makes you totally control what you want to check.

resource/dku-helpers.js
```diff
@@ -0,0 +1,93 @@
...
+            throw new Error("Mandatory column '" + param.name + "' not specified.");
+        }
+    }
+}
+
+function checkWebAppParameters(webAppConfig, webAppDesc) {
+    if (webAppDesc.topBarParams) {
+        webAppDesc.topBarParams.forEach(p => {checkMandatoryParameters(p, webAppConfig)});
```
HenriChabert: nitpick: When method has only one instruction, you do not need to use `{...}`.

resource/dku-helpers.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+function quadraticScalingFunction(min, max, total, value) {
+    if (max === min) {
+        return 0;
+    } else {
+        var scale = 1 / (max - min);
+        return Math.pow(Math.max(0,(value - min)*scale), 2);
+    }
```
HenriChabert: You do not need to use `else`. If condition in if verified, the function will return and the following instructions won't be executed.
```suggestion
    if (max === min) return 0;
    var scale = 1 / (max - min);
    return Math.pow(Math.max(0,(value - min)*scale), 2);
```

resource/dku-helpers.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+
+dataiku.webappBackend = (function() {
+    function getUrl(path) {
+        return dataiku.getWebAppBackendUrl(path);
+    }
+
+    function dkuDisplayError(error) {
+        console.warn("backend error: ", error);
```
HenriChabert: Why using `console.warn()` insted of `console.error()` ?

resource/dku-helpers.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    function get(path, args={}, displayErrors=true) {
+        return fetch(getUrl(path) + '?' + $.param(args), {
```
HenriChabert:
```suggestion
        return fetch(`${getUrl(path)}?${$.param(args)}`, {
```
nitpick: I think that this syntax is most of the time clearer

resource/dku-helpers.js
```diff
@@ -0,0 +1,93 @@
...
+            headers: {
+                'Accept': 'application/json',
+                'Content-Type': 'application/json'
+            }
+        })
+            .then(response => {
+                if (response.status == 502) {
+                    throw Error("Webapp backend not started");
+                } else if (!response.ok) {
+                    response.text().then(text => dataiku.webappMessages.displayFatalError(`Backend error:\n${text}.\nCheck backend log for more information.`))
```
HenriChabert, line 58: Indeed for the backend start. Why don' t you put it in app.js?

resource/leftBarParamsTemplate.html
```diff
@@ -0,0 +1,137 @@
...
+                <input type="radio" ng-model="config.maptile" value="mapbox/v13">
+                MapBox V13
+        </label>
+
+
+</form>
+
+
+<!--
+&lt;!&ndash;TODO: Potentially leave only this &ndash;&gt;
```
HenriChabert: I have tried before to just copy/paste the code of DSS and it has never worked. I think it is better to understand and to replicate only what you need.
About the dku-slider, it could be a great idea. Have you already tried using it ?
About the rest, it would indeed be good to replicate DKU style but it is already pretty nice I think.

webapps/density-custom-chart/webapp.json
```diff
@@ -0,0 +1,39 @@
+{
+    "meta": {
+        "label": "Geospatial density chart",
+        "description": "Custom chart to visualize datasets as graphs. These datasets must have source and target columns.",
```
HenriChabert: Did you forget to modify it when you used the code of Stan?

webapps/density-custom-chart/webapp.json
```diff
@@ -0,0 +1,39 @@
...
+        "icon": "icon-link"
```
HenriChabert: Would use a different icon.

webapps/density-custom-chart/body.html
```diff
@@ -0,0 +1,17 @@
+<script src="/plugins/geospatial-toolkit/resource/dku-helpers.js"></script>
+
+<script src="https://d3js.org/d3.v4.js"></script>
+<script src="https://d3js.org/d3-geo-projection.v2.min.js"></script>
+<script src="https://d3js.org/d3-quadtree.v1.min.js"></script>
+
+<link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css" integrity="sha512-xodZBNTC5n17Xt2atTPuE1HxjVMSvLVW9ocqUKLsCC5CXdbqCmblAshOMAS6/keqq/sMZMZ19scR4PsZChSR7A==" crossorigin=""/>
+<script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js" integrity="sha512-XQoYMqMTK8LvdxXYG3nZ448hOEQiglfqkJs1NOQV44cWnUrBc8PkAOcXy20w0vlaXaVUearIOBhiXZ5V3ynxwA==" crossorigin=""></script>
+<script src="/plugins/geospatial-toolkit/resource/libjs/leaflet-heat.js"></script>
+<script type="text/javascript"></script>
```
HenriChabert: Empty script

webapps/density-custom-chart/backend.py
```diff
@@ -0,0 +1,56 @@
...
+
+        #
+        logger.info("Filters {}".format(filters))
+        # TODO: Input sanitizer
+        dataset_name = config.get('dataset_name')
+        geopoint_column_name = config.get('geopoint_column_name')
+        # Fetch dataset from name
+        # TODO: Secure call to API
+        df = dataiku.Dataset(dataset_name).get_dataframe(limit=100000)
+        if len(filters) > 0:  # apply filters to dataframe
```
HenriChabert:
```suggestion
        if filters:  # apply filters to dataframe
```

webapps/density-custom-chart/backend.py
```diff
@@ -0,0 +1,56 @@
...
+            df = filter_dataframe(df, filters)
+
+        tooltip = ['A sample tooltip' for i in range(len(df))]
+        coordinates = list(df[geopoint_column_name].values)
+        long = [eval(i[6:-1].split(' ')[0]) for i in coordinates]
+        lat = [eval(i[6:-1].split(' ')[1]) for i in coordinates]
```
HenriChabert:
```suggestion
        long_, lat = zip(*[list(map(float, re.findall("\d+\.\d+", x))) for x in l])
```
Regex is more self explainatory I find but both are good, it is only to present you another way to do. Good practice also to use `long_` as long exists in Python builtin methods

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,356 @@
...
+    var _position;
+
+    var _intensity;
+    var _radius;
+    var _gradient;
+
+    Object.defineProperty(this, 'foo', {
+        get: function(){ return _foo;},
+        set: function(val){ _foo = val;}
+    });
```
HenriChabert: I guess this is a test?

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,356 @@
...
+    Object.defineProperty(this, 'mapPointer', {
+        get: function(){ return _mapPointer;}
+    });
+
+    this.do = function(val){
+        _foo += val;
+    }
+
+    this.initialiseMap = function(){
+        _mapPointer = L.map(_mapId).setView([48.8444529, 2.3718228], 5);
```
HenriChabert: Use constants to set DEFAULT_CENTER and DEFAULT_ZOOM

HenriChabert (changes requested):
Nice work !
I did not add reviews on the files that we checked together by Zoom. I'll look at it once you changed and tried what we discussed of :)

resource/dku-helpers.js
tdesfont: OK replaced mainly by let

js/leftBarParamsModule.js
tdesfont: Good idea, will do

python-lib/dku_filtering/filtering.py
tdesfont: Agreed

resource/dku-helpers.js
tdesfont: Noted

resource/dku-helpers.js
tdesfont: What should we put in app.js ?

webapps/density-custom-chart/webapp.json
tdesfont: Yes thanks

webapps/density-custom-chart/body.html
tdesfont: Removed

webapps/density-custom-chart/app.js
tdesfont: Removed

resource/dku-helpers.js
tdesfont: Will do

PR #2, opened 2021-01-11
[trade area processor] main dev branch for the trade area processor built with Gradle

tdesfont:
The aim of this feature development is to provide a geometry generator around an input center geopoint. Given a Well Known Text column of geopoints POINT(..., ...) in DSS, provide a processor (Prepare recipe step) that will take as input the distance (radius if circular area or width/height if rectangular area) and output the Well Known Text MultiPolygon defining the circular/rectangular area around the input points. User will then be able to visualise the trade area as a standard geometry in DSS and later perform a geo join of points in the created polygons. The build is done in Gradle (inspired from KML plugin) and the processor is in Java (inspired from Numeric Base Conversion plugin). To work inside DSS, one needs to build the jar in dist and then restart DSS.


Makefile
```diff
@@ -1,47 +1,9 @@
...
-plugin_id=`cat plugin.json | python -c "import sys, json; print(str(json.load(sys.stdin)['id']).replace('/',''))"`
-plugin_version=`cat plugin.json | python -c "import sys, json; print(str(json.load(sys.stdin)['version']).replace('/',''))"`
-archive_file_name="dss-plugin-${plugin_id}-${plugin_version}.zip"
-remote_url=`git config --get remote.origin.url`
-last_commit_id=`git rev-parse HEAD`
-
+PLUGIN_VERSION=0.1.0
+PLUGIN_ID=geospatial-toolkit
 
 plugin:
```
MayeulRousselet, line 4: don't forget to add the command to run the unit-tests with gradle

build.gradle
```diff
@@ -0,0 +1,116 @@
...
+task copyDependencies(type: Copy) {
+    from configurations.runtimeOnlyDeps
+    from configurations.remoteCompileDeps
+    into 'java-lib'
+}
+
+sourceSets {
+    main {
+        java {
+            srcDirs = ['src/main/java']
```
MayeulRousselet: Add an intermediary directory (java-processors)

js/processor.js
```diff
@@ -0,0 +1,33 @@
+(function() {
```
MayeulRousselet, line 1: What is the use of this file? It still references numerical conversion

plugin.json
```diff
@@ -10,8 +10,19 @@
...
+            "Geospatial"
         ],
         "supportLevel": "NOT_SUPPORTED"
-    }
+    },
+    "javaPreparationProcessors": [
+        {
+            "id": "tradearea",
+            "meta": {
+                "label": "Base conversion",
```
MayeulRousselet: To update

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+    double radius;
+
+    public RectangleAreaGenerator(double width, double height){
+        this.width = width;
+        this.height = height;
+        this.radius = Math.sqrt(Math.pow(width, 2)+Math.pow(height, 2));
+    }
+
+    public String generateArea(MyGeoPoint center) {
+        System.out.println("Generating an Rectangular area");
```
MayeulRousselet: Don't log with System.out.println but use a logger as you did in KML, and log variable data (width and height)

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+        double startLat = center.latitude;
+        double startLong = center.longitude;
+        MyGeoPoint initGeoPoint = null;
+        MyGeoPoint tmpGeoPoint;
+        StringBuilder str = new StringBuilder();
+        str.append("MULTIPOLYGON((");
```
MayeulRousselet: Why a multipolygon?

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+        double alpha = Math.atan(this.height/this.width);
```
MayeulRousselet: give a better name

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+        double[] angles = {90-alpha, 90+alpha, 270-alpha, 270+alpha};
```
MayeulRousselet: Which angles are these?

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public RectangleAreaGenerator(double width, double height){
+        this.width = width;
+        this.height = height;
+        this.radius = Math.sqrt(Math.pow(width, 2)+Math.pow(height, 2));
+    }
+
+    public String generateArea(MyGeoPoint center) {
+        System.out.println("Generating an Rectangular area");
+        double startLat = center.latitude;
```
MayeulRousselet: bad name, keep center.latitude as it is way clearer

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+        double startLong = center.longitude;
```
MayeulRousselet: same here

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+        MyGeoPoint initGeoPoint = null;
+        MyGeoPoint tmpGeoPoint;
```
MayeulRousselet: Don't need to be declared out of the loop

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,75 @@
...
+        StringBuilder str = new StringBuilder();
+        str.append("MULTIPOLYGON((");
+        double alpha = Math.atan(this.height/this.width);
+        double[] angles = {90-alpha, 90+alpha, 270-alpha, 270+alpha};
+        double angle;
```
MayeulRousselet: don't need to declare the variable outta loop

MayeulRousselet (changes requested):

js/processor.js
MayeulRousselet: Oooooook I found the usage of this file.
Its main purpose is to show the description of the processor in the shaker, as shown here
Capture d’écran 2021-01-11 à 4 19 39 PM

Makefile
tdesfont: Done in [414d8f64373baf4010cf325f7e71fbc8a9e9be13](url)

tdesfont: Done

build.gradle
tdesfont: Done

plugin.json
tdesfont: Done

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Changed to POLYGON

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Set to diagonalAngle

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Detailed in the wiki https://analytics.dataiku.com/projects/RDWIKI/wiki/1092/Plugin:%20GeoSpatial%20Toolkit/edit

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Put it inside

src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        @Override
+        public String getHelp() {
+            return "This processor performs trade area creation." +
+                    "\n \n" +
+                    "# Input column\n" +
+                    "Contains the geopoints on which the trade area is centered\n \n" +
+                    "# Output column\n" +
+                    "Contains created trade areas as WKT polygons (String column)\n\n"+
```
MayeulRousselet:
```suggestion
                    "Contains created trade areas in the WKT format\n\n"+
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
```diff
@@ -0,0 +1,43 @@
+package com.dataiku.dip.plugins.tradearea;
+
+import com.dataiku.dip.shaker.types.GeoPoint;
+
+public class GeoUtils {
+
+    // radians/degrees conversion
```
MayeulRousselet: Field comments must be in the following format:
```suggestion
    /**
     * radians/degrees conversion
     */
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
```diff
@@ -0,0 +1,43 @@
...
+    static double degreesToRadians = Math.PI/180;
+    static double radiansToDegrees = 180/Math.PI;
+    // earth radius in km
```
MayeulRousselet: Same here

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
```diff
@@ -0,0 +1,43 @@
...
+    static double earthRadius = 6378.137;
+
+    public static GeoPoint.Coords computeDestinationPoint(double latitude, double longitude, double bearing, double distance){
```
MayeulRousselet, line 26: Nice comment but not in the expected java format. See https://www.oracle.com/fr/technical-resources/articles/java/javadoc-tool.html for the specs. Anyway, the comment should be moved out of the method and follow this format:
```suggestion
/**
        Given a starting point latitude and longitude, compute latitude and longitude of the unique final point defined as the
        projection of the starting point on a circle of radius equal to `distance` (km) and with an angle equal to `bearing` (degrees).
        A bearing angle of 0 will result to output point being at North of starting point
        A bearing angle of 90 will result to output point being at West of starting point ...
        @param latitude: Latitude angle of the starting point in degrees
        @param  longitude: Longitude angle of the starting point in degrees
        @param bearing: Bearing angle in degrees (0 is north, 90 is west ... ), angle can be any real number
        @param distance: Distance in kilometers
       @return Coordinates of the resulting point
         */
    public static GeoPoint.Coords computeDestinationPoint(double latitude, double longitude, double bearing, double distance){
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
+package com.dataiku.dip.plugins.tradearea;
+
+import com.dataiku.dip.shaker.types.GeoPoint;
+import com.dataiku.dip.utils.DKULogger;
+
+abstract class AreaGenerator {
```
MayeulRousselet, line 9:
```suggestion
/**
    An abstract class expected by the TradeAreaProcessor to generate the polygons
     */
abstract class AreaGenerator {
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+    /*
+    An abstract class expected by the TradeAreaProcessor to generate the polygons
+     */
+
+    public abstract String generateArea(GeoPoint.Coords coords);
+
+}
+
+class RectangleAreaGenerator extends AreaGenerator {
```
MayeulRousselet, line 57:
```suggestion
/**
    Generates a rectangular area centered on an input geopoint coordinates.
    The parameters width, height, radius are expected to be in kilometers.
    If using miles in the processor, those distances must be converted to miles before.
     */
class RectangleAreaGenerator extends AreaGenerator {
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+            // Skip variable assignation if bad input
+        } else {
+            this.width = width;
+            this.height = height;
+            // Compute the radius based on half the width and height
+            this.radius = Math.sqrt(Math.pow(width/2, 2)+Math.pow(height/2, 2));
+        }
+    }
+
+    public String generateArea(GeoPoint.Coords center) {
```
MayeulRousselet, line 91:
```suggestion
/**
        Generate a rectangular trade area centered on an input geopoint coordonates.
        @param center : Coordinates of the center geopoint
        @return The rectangular trade area expressed as a WKT polygon (string)
            example: `POLYGON((long1 lat1,long2 lat2, ...))`
         */
    public String generateArea(GeoPoint.Coords center) {
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+        if (this.radius<=0 || this.width<=0 || this.height<=0){
+            logger.info("Detected invalid input parameter. Distance input parameters should be greater than zero.");
+            return null;
+        }
+        GeoPoint.Coords initCoords = null;
+        // Declaration of the final result
+        StringBuilder str = new StringBuilder();
+        str.append("POLYGON((");
+        // Computation of the diagonal angle of the rectangle (Must be in degree and Math.atan is radian)
+        double diagonalAngle = Math.atan(this.height/this.width)*180/Math.PI;
```
MayeulRousselet: As this variable do not depends on the method parameter, declared it as a field and iinit it in the constructor.

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+        double[] angles = {90-diagonalAngle, 90+diagonalAngle, 270-diagonalAngle, 270+diagonalAngle};
+        // Compute the latitude longitude of the four corners and fill the polygon String
+        for (int i = 0; i < angles.length; i++){
+            double angle;
+            angle = angles[i];
```
MayeulRousselet, line 106:
```suggestion
            double angle = angles[i];
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+            GeoPoint.Coords tmpCoords;
+            tmpCoords = GeoUtils.computeDestinationPoint(center.latitude, center.longitude, angle, this.radius);
```
MayeulRousselet:
```suggestion
            GeoPoint.Coords tmpCoords = GeoUtils.computeDestinationPoint(center.latitude, center.longitude, angle, this.radius);
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+            str.append(tmpCoords.longitude + " " + tmpCoords.latitude + ",");
```
MayeulRousselet: Avoid temporary string creation by using multiple append.
```suggestion
str.append(tmpCoords.longitude);
str.append(" ");
str.append(tmpCoords.latitude);
str.append(",");
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        GeoPoint.Coords initCoords = null;
+        // Declaration of the final result
+        StringBuilder str = new StringBuilder();
+        str.append("POLYGON((");
+        // Computation of the diagonal angle of the rectangle (Must be in degree and Math.atan is radian)
+        double diagonalAngle = Math.atan(this.height/this.width)*180/Math.PI;
+        double[] angles = {90-diagonalAngle, 90+diagonalAngle, 270-diagonalAngle, 270+diagonalAngle};
+        // Compute the latitude longitude of the four corners and fill the polygon String
+        for (int i = 0; i < angles.length; i++){
```
MayeulRousselet: Start at i=1 and handle the initCoord specifically

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+            angle = angles[i];
+            GeoPoint.Coords tmpCoords;
+            tmpCoords = GeoUtils.computeDestinationPoint(center.latitude, center.longitude, angle, this.radius);
+            str.append(tmpCoords.longitude + " " + tmpCoords.latitude + ",");
+            if (i==0){
+                initCoords = tmpCoords;
+            }
+        }
+        // Close the polygon by adding initial geopoint
+        str.append(initCoords.longitude + " " + initCoords.latitude);
```
MayeulRousselet: Remove the "+"

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+        str.append("))");
+        return str.toString();
+    }
+}
+
+class CircleAreaGenerator extends AreaGenerator {
+    /*
```
MayeulRousselet: Use a class comment

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+
+    double radius;
+
+    public CircleAreaGenerator(double radius){
+        this.radius = radius;
+    }
+
+    public String generateArea(GeoPoint.Coords center) {
+        /*
```
MayeulRousselet: Use the good method foramt

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+        Generate an almost circular area approximated with 12 points
+
+        Input:
+            GeoPoint.Coords center: The center of the trade area as a geospatial instance coordinates
+        Output:
+            The circular trade area expressed as a WKT POLYGON (String)
+            example: `POLYGON((long1 lat1,long2 lat2, ...))`
+         */
+        if (this.radius<=0){
+            logger.info("Detected invalid radius as input. Radius parameter should be greater than zero.");
```
MayeulRousselet: Log the radius value

MayeulRousselet: Log the radius value

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+            return null;
+        }
+        GeoPoint.Coords initCoords = null;
+        StringBuilder str = new StringBuilder();
+        str.append("POLYGON((");
+        // Compute the points on circle, angle step is set to 30 degrees as there are 12 points
+        for (int i = 0; i < 12; i++){
```
MayeulRousselet: Factorize with the method for the rectangle.

MayeulRousselet: Create a constant for 12:
```suggestion
class CircleAreaGenerator extends AreaGenerator {
       static final int NB_OF_EDGES = 360;
       ....
       for (int i = 0; i < NB_OF_EDGES; i++){
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
```diff
@@ -0,0 +1,43 @@
...
+            bearing: Bearing angle in degrees (0 is north, 90 is west ... ), angle can be any real number
+            distance: Distance in kilometers
+
+        Output:
+            Coordinates of the resulting point
+
+         */
+        latitude *= degreesToRadians;
+        longitude *= degreesToRadians;
+        bearing *= degreesToRadians;
```
MayeulRousselet: Use radians for bearing and not degrees

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Sets;
+
+public class TradeAreaProcessor extends SingleInputSingleOutputRowProcessor implements Processor {
+
+    public static class Parameter implements StepParams {
+        private static final long serialVersionUID = -1;
+        public String inputColumn;
+        public String outputColumn;
+        // Custom parameters declaration of the processor
```
MayeulRousselet:
```suggestion
        /** 
          * Custom parameters declaration of the processor
          */
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+        public UnitMode unitMode;
+        public ShapeMode shapeMode;
+        public double radius;
+        public double width;
+        public double height;
+
+        @Override
+        public void validate() throws IllegalArgumentException {
+            TradeAreaProcessor.newGenerator(unitMode, shapeMode, radius, height, width);
```
MayeulRousselet: It will be better to implement a static validate method on TradeAreaProcessor

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+            outputColumn = getCf().column(params.outputColumn, ProcessorRole.OUTPUT_COLUMN);
+        } else {
+            outputColumn = cd;
+        }
+        generator = newGenerator(params.unitMode, params.shapeMode, params.radius, params.height, params.width);
+    }
+
+    @VisibleForTesting
+    static AreaGenerator newGenerator(UnitMode unitMode, ShapeMode shapeMode, double radius, double height, double width) {
+        logger.info("Instantiated new generator with argument: width="+width+" height="+height);
```
MayeulRousselet: Better to use infoV to log variable arguments + log the radius
```suggestion
        logger.infoV("Instantiated new generator with argument: width=%f height=%f", width, height);
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+        // Conversion factor from miles to km to enforce km as input
+        double milesToKm = 1.6093444978925633;
```
MayeulRousselet: declares it as a class constant 
```suggestion
      static final double MILES_TO_KM = 1.6093444978925633;
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+        // Expected input distances in the area generators are in km
+        // If in miles, convert to km
+        logger.info("Using miles as standard unit for distances");
```
MayeulRousselet: I think we use km as standard instead of miles

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/CircleAreaGeneratorTest.java
```diff
@@ -0,0 +1,18 @@
+package com.dataiku.dip.plugins.tradearea;
+
+import static org.junit.jupiter.api.Assertions.*;
+import org.junit.jupiter.api.Test;
+
+import com.dataiku.dip.shaker.types.GeoPoint;
+
+class CircleAreaGeneratorTest {
```
MayeulRousselet, line 8: Add tests for invalid input etc, as you did for the Rectangle

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    public void testInvalidInput() {
+        double width = 8;
+        double height = 5;
+        String actualResult = null;
+        RectangleAreaGenerator generator = new RectangleAreaGenerator(width, height);
+        String str = "Bad input";
+        GeoPoint.Coords coords = GeoPoint.convert(str);
```
MayeulRousselet: GeoPoint.convert will return null for "Bad Input", therefore you don't test your code as you never enter in the if.
better to test generator.generateArea with a null input.

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    public void testInvalidFormat() {
+        double width = 8;
+        double height = 5;
+        String actualResult = null;
+        RectangleAreaGenerator generator = new RectangleAreaGenerator(width, height);
+        String str = "POINT(-73.9723, 40.64749)";
+        GeoPoint.Coords coords = GeoPoint.convert(str);
```
MayeulRousselet: Here you only test for GeoCoord.convert, which must be tested beforehand

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Test
+    public void testInvalidRange() {
+        double width = 8;
+        double height = 5;
+        String actualResult = null;
+        RectangleAreaGenerator generator = new RectangleAreaGenerator(width, height);
+        String str = "POINT(-73.9723 1000)";
+        GeoPoint.Coords coords = GeoPoint.convert(str);
+        System.out.println(coords);
```
MayeulRousselet: Remove the System.out

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
```diff
@@ -0,0 +1,82 @@
...
+        if (coords != null){
```
MayeulRousselet: Coords il never null, remove the test

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test
+    public void testNullDistances() {
+        double width = 0;
+        double height = 0;
+        String actualResult = null;
+        RectangleAreaGenerator generator = new RectangleAreaGenerator(width, height);
+        String str = "POINT(-73.9723 40.64749)";
+        GeoPoint.Coords coords = GeoPoint.convert(str);
+        System.out.println(coords);
+        if (coords != null){
```
MayeulRousselet: Coords will not be null, remove the if and remove the System.out.println

MayeulRousselet (changes requested):

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok now

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok now

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok now

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok

tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Ok now

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Did you decide to do that at the end ?

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
tdesfont: It might be a bit strange to have latitude, longitude in degrees and bearing in radians

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/GeoUtils.java
MayeulRousselet: Not necessarily if the variable name reflects it

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Not sure to understand, will commit the following
```
    static void validate(UnitMode unitMode, ShapeMode shapeMode, double radius, double height, double width){
        AreaGenerator areaGenerator = newGenerator(unitMode, shapeMode, radius, height, width);
    }
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Done

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/CircleAreaGeneratorTest.java
tdesfont: Done

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
tdesfont: Right ! Fixed

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
tdesfont: Should I remove any tests that test GeoCoord.convert ?

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
tdesfont: Done

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
tdesfont: Done

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -177,25 +184,34 @@ public void init() throws Exception {
...
             case CIRCLE:
+                logger.info("Instantiating a Circular Area Generator.");
                 return new CircleAreaGenerator(radius);
             default:
                 throw new IllegalArgumentException("Invalid processing mode: " + shapeMode);
             }
     }
+
+    static void validate(UnitMode unitMode, ShapeMode shapeMode, double radius, double height, double width){
+        AreaGenerator areaGenerator = newGenerator(unitMode, shapeMode, radius, height, width);
```
MayeulRousselet: You should have a specific method to validate without creating a new AreaGenerator.
You should check for shapeMode, unitMode, radius > 0 (if cricle) and height & witdh (if rectancle).

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -98,28 +99,26 @@ public CircleAreaGenerator(double radius){
      * @return The circular trade area expressed as a WKT POLYGON (String `POLYGON((long1 lat1,long2 lat2, ...))`)
      */
     public String generateArea(GeoPoint.Coords center) {
+        if (center == null){
+            return null;
+        }
 
         if (this.radius<=0){
```
MayeulRousselet: No need to check the radius if it was checked before in the validation, the value can't be changed.

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -45,35 +46,35 @@ public RectangleAreaGenerator(double width, double height){
      * @return The rectangular trade area expressed as a WKT polygon (string)
      */
     public String generateArea(GeoPoint.Coords center) {
+        if (center == null){
+            return null;
+        }
         if (this.radius<=0 || this.width<=0 || this.height<=0){
```
MayeulRousselet: No need to check for these value as it should be done by the validation.

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/PolygonBuilder.java
```diff
@@ -0,0 +1,26 @@
+package com.dataiku.dip.plugins.tradearea;
+
+import com.dataiku.dip.shaker.types.GeoPoint;
+
+public class PolygonBuilder {
+
+    StringBuilder wktPolygon;
```
MayeulRousselet: add final (the variable won't be reassigned after the constructor has been called)
```suggestion
    final StringBuilder wktPolygon;
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/PolygonBuilder.java
```diff
@@ -0,0 +1,26 @@
...
+    GeoPoint.Coords firstPoint;
```
MayeulRousselet: Add final

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/PolygonBuilder.java
```diff
@@ -0,0 +1,26 @@
+package com.dataiku.dip.plugins.tradearea;
+
+import com.dataiku.dip.shaker.types.GeoPoint;
+
+public class PolygonBuilder {
```
MayeulRousselet, line 12: If there is no call to lineTo, the resulting polygon will be bogus (the 2 same points without a comma between the two).
Either document in the class comment that a lineTo call is mandatory, or handle the case.

MayeulRousselet (changes requested):

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Done in 371c918

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: done in e0f3eaf

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: done in e0f3eaf

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/PolygonBuilder.java
tdesfont: Done in 4a72e51

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/PolygonBuilder.java
tdesfont: Done in f479de8

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/PolygonBuilder.java
tdesfont: Done in 72e5797

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+    public static void checkInputParams(TradeAreaProcessor.UnitMode unitMode, TradeAreaProcessor.ShapeMode shapeMode,
+                                           double radius, double height, double width){
+
+        switch (shapeMode) {
+        case RECTANGLE:
+            if (width <= 0 || height <= 0) {
+                logger.info("Rectangle Area Generator: Received invalid parameters as input.");
+                logger.infoV("Got width= {}", width);
+                logger.infoV("Got height= {}", height);
+                throw new IllegalArgumentException(
```
MayeulRousselet: I am not sure about the throw as it will result in a ugly message on the front: See for example:
Capture d’écran 2021-02-08 à 9 51 35 PM

I've dig in the DIP code to look for validation in processor and very few of them do it.
I think we should only check for shapeMode and UnitMode and raise an exception if they are not known, and silently do nothing if radius or width or height are <=0.

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+         */
+        public UnitMode unitMode;
+        public ShapeMode shapeMode;
+        public double radius;
+        public double width;
+        public double height;
+
+        @Override
+        public void validate() throws IllegalArgumentException {
+            TradeAreaProcessor.validate(unitMode, shapeMode, radius, height, width);
```
MayeulRousselet: Directly call AreaGenerator. checkInputParams here and remove TradeAreaProcessor.validate

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Generates a rectangular area centered on an input geopoint coordinates.
+ * The parameters width, height, radius are expected to be in kilometers.
+ * If using miles in the processor, those distances must be converted to miles before.
+ */
+class RectangleAreaGenerator extends AreaGenerator {
+
+    double width;
```
MayeulRousselet: declare each field as final

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+     * @param center: Coordinates of the center geopoint
+     * @return The rectangular trade area expressed as a WKT polygon (string)
+     */
+    public String generateArea(GeoPoint.Coords center) {
+        if (center == null){
+            return null;
+        }
+
+        // Declaration of the final result
+        StringBuilder str = new StringBuilder();
```
MayeulRousselet: str is not used anymore, remove it

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Generates a circular area centered on an input geopoint coordinates.
+ * The parameter radius are expected to be in kilometers.
+ * If using miles in the processor, those distances must be converted to miles before.
+ */
+class CircleAreaGenerator extends AreaGenerator {
+    static final int NB_OF_EDGES = 12;
+    static final double ANGLE_STEP = 2*Math.PI/NB_OF_EDGES;
+
+    double radius;
```
MayeulRousselet: declare final

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    /**
+     * Generate a rectangular trade area centered on an input geopoint coordonates.
+     * @param center: Coordinates of the center geopoint
+     * @return The rectangular trade area expressed as a WKT polygon (string)
+     */
+    public String generateArea(GeoPoint.Coords center) {
+        if (center == null){
```
MayeulRousselet:
```suggestion
        if (center == null || diagonalAngle==0){
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+        this.radius = radius;
+    }
+
+    /**
+     * Generate an almost circular area approximated with 12 points
+     * @param center: The center of the trade area as a geospatial instance coordinates
+     * @return The circular trade area expressed as a WKT POLYGON (String `POLYGON((long1 lat1,long2 lat2, ...))`)
+     */
+    public String generateArea(GeoPoint.Coords center) {
+        if (center == null){
```
MayeulRousselet:
```suggestion
        if (center == null || radius==0){
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+    double radius;
+    double diagonalAngle;
+
+    private static DKULogger logger = DKULogger.getLogger("dku");
+
+    public RectangleAreaGenerator(double width, double height){
+        if (width <= 0 || height <= 0){
+            logger.info("Rectangle Area Generator: Received invalid parameters as input.");
+            logger.infoV("Got width= {}", width);
+            logger.infoV("Got height= {}", height);
```
MayeulRousselet: init this.height, this.width and this.diagonalAngle to 0

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+ * If using miles in the processor, those distances must be converted to miles before.
+ */
+class CircleAreaGenerator extends AreaGenerator {
+    static final int NB_OF_EDGES = 12;
+    static final double ANGLE_STEP = 2*Math.PI/NB_OF_EDGES;
+
+    double radius;
+
+    public CircleAreaGenerator(double radius){
+        this.radius = radius;
```
MayeulRousselet:
```suggestion
        this.radius = Math.max(0, radius);
```

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+                        "Circular Trade Area Processor Failed.\n"+
+                        "The radius of the trade area must be strictly greater than 0.\n"+
+                        "Received radius value is equal to "+ radius);
+            }
+            break;
+        default:
+            throw new IllegalArgumentException("Invalid processing mode: " + shapeMode);
+        }
+    };
+
```
MayeulRousselet, line 49: Check for UnitMode too

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+                return new RectangleAreaGenerator(width, height);
+            case CIRCLE:
+                logger.info("Instantiating a Circular Area Generator.");
+                return new CircleAreaGenerator(radius);
+            default:
+                throw new IllegalArgumentException("Invalid processing mode: " + shapeMode);
+            }
+    }
+
+    static void validate(UnitMode unitMode, ShapeMode shapeMode, double radius, double height, double width){
```
MayeulRousselet: TO remove

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/CircleAreaGeneratorTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Test
+    public void testInvalidInput() {
+        double radius = 9;
+        CircleAreaGenerator generator = new CircleAreaGenerator(radius);
+        String actualResult = generator.generateArea(null);
+        assertNull(actualResult);
+    }
+
+    @Test
+    public void testInvalidFormat() {
```
MayeulRousselet: You don't have to test for that, It should be tested in DSS.

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/CircleAreaGeneratorTest.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    public void testInvalidRange() {
+        double radius = 8;
+        CircleAreaGenerator generator = new CircleAreaGenerator(radius);
+        String str = "POINT(-73.9723 1000)";
+        GeoPoint.Coords coords = GeoPoint.convert(str);
+        String actualResult = generator.generateArea(coords);
+        assertNotNull(actualResult);
```
MayeulRousselet: You should test against a real expected value.

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void testInvalidInput() {
+        double width = 8;
+        double height = 5;
+        RectangleAreaGenerator generator = new RectangleAreaGenerator(width, height);
+        String actualResult = generator.generateArea(null);
+        assertNull(actualResult);
+    }
+
+    @Test
+    public void testInvalidRange() {
```
MayeulRousselet: You don't test an invalid range here, you do the same test as in testStandardGenerator. Delete this test

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+        getProcessorOutput().lastRowEmitted();
+    }
+
+    @Override
+    public void init() throws Exception {
+        cd = getCf().column(params.inputColumn, ProcessorRole.INPUT_COLUMN);
+        if (!StringUtils.isBlank(params.outputColumn)) {
+            outputColumn = getCf().column(params.outputColumn, ProcessorRole.OUTPUT_COLUMN);
+        } else {
+            outputColumn = cd;
```
MayeulRousselet, line 185: There is a bug here.
Let say I have three columns in my dataset (lon, lat and location).
I choose lon as an input col, and no output column, then the value of the lon columns are wiped out.
Cf screenshot below:
Capture d’écran 2021-02-08 à 10 43 50 PM

MayeulRousselet (changes requested):

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
tdesfont: Done in 4fe5ecb

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: See 7a41321
Added ignore on invalid radius or width/height value, removed the throw error message for invalid width/height/radius

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Done in b225e00

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done in 8a82db6

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done in 8214dc5

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done in 3b270d4

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done in e30115b

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done in a8002c2

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done in 16c0519

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/AreaGenerator.java
tdesfont: Done

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Done

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/CircleAreaGeneratorTest.java
tdesfont: Done

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/CircleAreaGeneratorTest.java
tdesfont: Done in 2fe6e21

java-processors/src/test/java/com/dataiku/dip/plugins/tradearea/RectangleAreaGeneratorTest.java
tdesfont: Done in f9fef19

java-processors/src/main/java/com/dataiku/dip/plugins/tradearea/TradeAreaProcessor.java
tdesfont: Can't reproduce the bug.

tdesfont: This is the standard behaviour of native processor, in our case, if null values is returned and there is a value in the input column, we decide to leave the input column value in the result column.

Screenshot 2021-02-15 at 11 25 18

tdesfont: Fixed in 1c592f4.

clemencebic:
Here are a few things we could improve on the processor : 
- we already said we want to increase the number of edges for the circle (we are still using 12 edges)
- we should reverse the order of the "distance unit" and "shape" parameters (the distance unit parameter should be closer to radius en width parameter)
- for radius, width and height I would add the following indication : "radius (for circle shape)", 'width (for rectangle shape)' ...
- Can we prevent the user from putting negative values in the width, height and radius parameters ?  Or at least raise an error
- Is it possible to add warnings ? For example if shape=circle and radius=0 could we put a warning "A 0 radius will generate empty values)
- In the processor description, we should also mention the "radius", "width" and 'height' parameters
- UI and wording of the trade area : 
image
1. We should keep the same type of syntax : "Generate" instead of "Generating"
2. We should keep the same UI than the other processors, that is "Generate a trade area" in bold, and "merchant location" in red in my example

tdesfont:
- [x] Increase the number of edges for the circle (Set to 36)
- [x] Reverse the order of the "distance unit" and "shape" parameters
- [x] add the following indication : "radius (for circle shape)", 'width (for rectangle shape)' ...
- [x] Prevent the user from putting negative values in the width, height and radius parameters 
- [ ] Add warnings ? For example if shape=circle and radius=0 could we put a warning "A 0 radius will generate empty values)
- [x] In the processor description, we should also mention the "radius", "width" and 'height' parameters
- [x] UI and wording of the trade area: Keep the same type of syntax : "Generate" instead of "Generating"
- [ ]  UI and wording of the trade area: We should keep the same UI than the other processors, that is "Generate a trade area" in bold, and "merchant location" in red in my example

PR #7, opened 2021-02-08
[density chart] Main pull request (1/2) for the geospatial toolkit (Wait for other prs before merge)

tdesfont:
Custom chart that displays a density map.


js/leftBarParamsModule.js
```diff
@@ -0,0 +1,3 @@
+var app = angular.module('geospatial.density', []);
+
+app.controller('MyCustomFormController', ['$scope', function($scope) {}]);
```
HenriChabert: Maybe you can use a more explicit controller name, for example `DensityCustomChartController`. You also have to change it in leftBarParamsTemplate.html

plugin.json
```diff
@@ -3,15 +3,15 @@
...
     "meta": {
         "label": "Geospatial toolkit",
-        "category": "Geosptial",
+        "category": "Geospatial",
         "description": "Tools for geospatial data preparation and visualisation",
         "author": "Dataiku",
-        "icon": "icon-font-awesome-3.2.1",
+        "icon": "icon-globe",
         "licenseInfo": "Apache Software License",
         "url": "https://www.dataiku.com/product/plugins/your-plugin-id/",
```
HenriChabert: Change here the plugin id, it will be ready for release

plugin.json
```diff
@@ -3,15 +3,15 @@
...
         "tags": [
-            "Geosptial"
+            "Geospatial"
```
HenriChabert: You can also add `Geo` and `Vizualisation`

resource/leftBarParamsTemplate.html
```diff
@@ -0,0 +1,124 @@
...
+        <!-- Map BackGround Selector -->
+
+        <div class="mbot20" style="padding-bottom: 5px; margin-bottom: 0; overflow: hidden;">
+                <h2 class="settings-section-title foldable-title" id="qa_charts_map-background-submenu-header" ng-click="config.optionsFolds.mapTiles = !config.optionsFolds.mapTiles">
+                        <i class="icon-caret-right" ng-show="!config.optionsFolds.mapTiles" /><i class="icon-caret-down" ng-show="config.optionsFolds.mapTiles" />
+                        Map background
+                </h2>
+
+                <div ng-show="config.optionsFolds.mapTiles" class="options-zone" ng-controller="MapBackgroundPickerController">
+                        <form class="dkuform-chart-options">
```
HenriChabert: As discussed together, I investigated the problem when placing Color before MapBackground. It comes from the fact that you have two nested forms and so there are some unexpected behaviours. What you have to do is to replace the `form` tag by `ng-form` as AngularJS supportes nested tags.
Once it has been done, you will be able to place the sections in any order.

resource/leftBarParamsTemplate.html
```diff
@@ -0,0 +1,124 @@
...
+                                <div class="control-group">
+                                        <label class="control-label" dku-for="#map-background-picker button">Tiles to use</label>
+                                        <div class="controls">
+                                                <div class="palette-picker__list-zone" id="map-background-picker">
+                                                        <div class="flex oh" id="palette-picker">
```
HenriChabert: Change the div ID to make it unique

resource/leftBarParamsTemplate.html
```diff
@@ -0,0 +1,124 @@
+
+<form name="myForm" ng-controller="MyCustomFormController">
```
HenriChabert: Actually to make things cleaner, this form should also be renamed `ng-form`. However, I am wondering, where id you see that you had to put so many nested forms? When looking at the Scatter plot, there is not any form nested, I think this is something that could be the problem of many things as it implies unexpected behaviour (it is not allowed in HTML).

webapps/density-custom-chart/webapp.json
```diff
@@ -0,0 +1,39 @@
...
+    "hasBackend": "true",
+
+    "chart" : {
+        "datasetParamName": "dataset",
+        "leftBarLabel": "Chart parameters",
+        "leftBarParams": [],
+        "leftBarTemplate": "leftBarParamsTemplate.html",
+        "leftBarModule": "geospatial.density",
+        "topBar": "STD_FORM",
+        "topBarParams": [
```
HenriChabert, line 18: There are some differents with generated templates of the topbar and the ones in the Scatter plot (For example, the inputs are centered while there are left aligned in the Scatter plot). To make it as close as possible, you would need to create a custom top bar as well.

webapps/density-custom-chart/webapp.json
```diff
@@ -0,0 +1,39 @@
+{
+    "meta": {
+        "label": "Density map",
+        "description": "Generate a density map from geospatial data. Takes as input WKT points.",
```
HenriChabert, line 4: What is WKT?

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    z-index: 1500;
+    position:absolute;
+    top:0; bottom:0; left:0; right:0;
+    display:block;
+    background:transparent;
+}
+
+.geo-chart-style {
+
+}
```
HenriChabert: Empty property

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    border-top-left-radius: 12px;
+    border-top-right-radius: 12px;
+    border-bottom-left-radius: 12px;
+    border-bottom-right-radius: 12px;
+    padding-left: 15px;
+    padding-right: 15px;
+    padding-top: 5px;
+    padding-bottom: 5px;
+    font-size: 12px;
+
```
HenriChabert: Give me some Shadow there, love shadow 🤩
```suggestion
box-shadow: 0 3px 14px rgb(0 0 0 / 40%);
```

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    font-family: 'SourceSansPro';
+    font-weight: 500;
+    font-size: 14px;
+    color: #ffffff;
+    box-sizing: border-box;
+}
+
+.error-message-style {
+    z-index: 1500;
+    background-color: #28a9dd;
```
HenriChabert, line 47: You'd better use Dataiku vars
```suggestion
    background-color: var(--error-red);
```

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+}
+
+@keyframes spin {
+    0% { transform: rotate(0deg); }
+    100% { transform: rotate(360deg); }
+}
+
+.user-message-style {
+    z-index: 1500;
+    background-color: #28a9dd;
```
HenriChabert:
```suggestion
    background-color: var(--blue-darken-2);
```

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    background:transparent;
+}
+
+.warning-view-style {
+    z-index: 1500;
+    position:absolute;
+    top:0; bottom:0; left:0; right:0;
+    display:block;
+    background:transparent;
+}
```
HenriChabert: Unused

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    z-index: 1500;
+    background-color: #28a9dd;
+    position: absolute;
+    top: 0px;
+    left: -1px;
+    margin-top: 10px;
+    width: 100%;
+    text-align: center;
+    padding: 10px;
+    font-family: 'SourceSansPro';
```
HenriChabert: Unnecessary, already inheritate the font from the body

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    font-weight: 500;
+    font-size: 14px;
+    color: #ffffff;
```
HenriChabert:
```suggestion
    color: var(---white);
```

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    position: absolute;
+    top: 0px;
+    left: -1px;
+    margin-top: 10px;
+    width: 100%;
+    text-align: center;
+    padding: 10px;
+    font-family: 'SourceSansPro';
+    font-weight: 500;
+    font-size: 14px;
```
HenriChabert, line 56: Unnecessary, already inheritate the font from the body

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
```
HenriChabert, line 40: Unnecessary, already inheritate the font from the body

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    z-index: 1500;
+    background-color: #28a9dd;
+    position: absolute;
+    top: 0px;
+    left: -1px;
+    margin-top: 10px;
+    width: 100%;
+    text-align: center;
+    padding: 10px;
+    font-family: 'SourceSansPro';
```
HenriChabert: Unnecessary, already inheritate the font from the body

webapps/density-custom-chart/style.css
```diff
@@ -0,0 +1,100 @@
...
+    font-weight: 500;
+    font-size: 14px;
+    color: #ffffff;
```
HenriChabert:
```suggestion
    color: var(--white);
```

webapps/density-custom-chart/body.html
```diff
@@ -0,0 +1,28 @@
...
+<script src="https://d3js.org/d3-scale.v3.min.js"></script>
+<script src="https://d3js.org/colorbrewer.v1.min.js"></script>
+
+<link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css" integrity="sha512-xodZBNTC5n17Xt2atTPuE1HxjVMSvLVW9ocqUKLsCC5CXdbqCmblAshOMAS6/keqq/sMZMZ19scR4PsZChSR7A==" crossorigin=""/>
+<script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js" integrity="sha512-XQoYMqMTK8LvdxXYG3nZ448hOEQiglfqkJs1NOQV44cWnUrBc8PkAOcXy20w0vlaXaVUearIOBhiXZ5V3ynxwA==" crossorigin=""></script>
+<script src="/plugins/geospatial-toolkit/resource/libjs/leaflet-heat.js"></script>
+
+<div style="text-align: center">
+
+    <div id="error-warning-view" class="error-warning-view-style" style="width: 100%; height: 100%; background-color: white;">
```
HenriChabert: Always better to place the style into the css rather than inline

webapps/density-custom-chart/body.html
```diff
@@ -0,0 +1,28 @@
...
+        <div id="error-message" class="error-message-style"></div>
+        <img src="/plugins/geospatial-toolkit/resource/media/density-map.png" style="padding-top: 50px; width: 40%;">
```
HenriChabert: Same here, assign a class and place the css into the css stylesheet

webapps/density-custom-chart/body.html
```diff
@@ -0,0 +1,28 @@
...
+    </div>
+
+    <div id="tooltip"></div>
+    <div id="spinner"></div>
+    <div id="geo-chart" class="geo-chart-style"><div id="mapid" style="width: 100%; height: 100%;"></div></div>
```
HenriChabert: Idem

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+            configEvent.needCoreDataUpdate = false;
+        } else {
+            if (chartHandler.coreData.length !== 0){
+                chartHandler.render();
+            }
+        }
+
+        if (!configEvent.geopointColumnName){
+            console.log("Display warning");
+            $('#error-message').css("background-color", "#28a9dd");
```
HenriChabert: Rather than changing the color, couldn't you change the class? It would allow every styling to be put in the css stylesheet.

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+        chartHandler.radius = configEvent.radius;
+        chartHandler.gradient = convertPaletteToGradient(colorPalette);
+
+        if (configEvent.needCoreDataUpdate){
+            $("#spinner").show();
+            console.log("Request of a full backend recompute of the data");
+            updateCoreData(configEvent, chartHandler);
+            configEvent.needCoreDataUpdate = false;
+        } else {
+            if (chartHandler.coreData.length !== 0){
```
HenriChabert: The !== is superfluous

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+        configEvent.detailsColumnName = webAppConfig['details_column_name'];
+        configEvent.tooltipColumnName = eventData['uaTooltip'];
+        configEvent.filters = filters;
+        configEvent.geopointColumnName = webAppConfig['geopoint_column_name'];
+
+        let colorPalette = configEvent.colorPalette;
+        chartHandler.colorPalette = colorPalette;
+        chartHandler.intensity = configEvent.intensity;
+        chartHandler.radius = configEvent.radius;
+        chartHandler.gradient = convertPaletteToGradient(colorPalette);
```
HenriChabert: Maye you could put all that in a new function `setConfigEvent()`

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+                chartHandler.render();
+                if (configEvent.needMapCentering) {
+                    console.log("Trying to center map.");
+                    chartHandler.centerMap();
+                    configEvent.needMapCentering = false;
+                }
+                chartHandler.initialised = true;
+                document.getElementById("error-message").text = "";
+                document.getElementById("error-warning-view").style.display = "none";
+                document.getElementById("spinner").style.display = "none";
```
HenriChabert: - To would be better to write a function `showSpinner()` and `hideSpinner()` and sae for `showErrorMessage` and `hideErrorMessage` as you repeat this code quite often.
- Moreover, it would be a little bit cleaner if you either use jQuery or Vanilla js for selection (Here, maybe you can use `$("spinner").hide()`

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+            } else {
+                console.log("Received no data");
+                $('#error-message').css("background-color", "red");
```
HenriChabert: Use a error class rather than inline css

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+                dataiku.webappMessages.displayFatalError(errorMessage);
+                document.getElementById("spinner").style.display = "none";
+            }
+        }).catch(error => {
+            console.error("Caught error:", error);
+            $('#error-message').css("background-color", "red");
+            dataiku.webappMessages.displayFatalError(error);
+            document.getElementById("spinner").style.display = "none";
```
HenriChabert: DRY: Place this in a function with errorMessage as a parameter

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+chartHandler.initialiseMap();
+chartHandler.addLeafletLayer();
+chartHandler.addMousePosition();
+chartHandler.addScatterPlotLayer();
+chartHandler.tooltip = tooltip;
+chartHandler.addSearchTrigger();
+chartHandler.addUpdateEvent();
+
+// Variables specific to the custom web app in DSS
+let webAppDesc = dataiku.getWebAppDesc()['chart'];
```
HenriChabert: Unused in file

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+    })
+        .then(function(data){
+            console.log("Finished query of server to fetch core data");
+            chartHandler.clear();
+            console.log("Received data:", data);
+            if (data.length !== 0){
+                let tempGeopoints = [];
+                for (let i = 0; i < data.length; i++) {
+                    tempGeopoints.push([data[i]['lat'], data[i]['long'], data[i]['detail'], data[i]['tooltip']]);
+                }
```
HenriChabert: Simplification
```suggestion
                let tempGeopoints = data.map(d => [d.lat, d.long, d.detail, d.tooltip]);
```

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+/**
+ * Main request to backend when data needs to be updated. (Non-blocking and potentially long)
+ * @param configEvent : The config object used to deduce actions from user parameters
+ * @param chartHandler : The chart handler object used to interact with map and layers in front end
+ */
+function updateCoreData(configEvent, chartHandler) {
+
+    console.log("Starting query of server to fetch core data ....");
```
HenriChabert, line 53: I would personally show the spinner inside this function rather as it is clearer for a code reader

webapps/density-custom-chart/app.js
```diff
@@ -0,0 +1,133 @@
...
+ *
+ *
+ */
+
+const errorMessage = "No geodata column, please select a valid geodata column";
+
+// Declare tooltip div that will be used for tooltip display
+let tooltip = d3.select("#tooltip") // tooltip for the information of the point
+    .append("div")
+    .attr("class", "tooltip");
```
HenriChabert: Is this div really useful? When removing it and hard coding the "tooltip" class on the parent div, it seems to work but maybe I missed something

resource/gradient-color-scale.js
```diff
@@ -0,0 +1,58 @@
...
+
+
+function findId(data, idToLookFor) {
+    var categoryArray = data.palettes;
+    for (var i = 0; i < categoryArray.length; i++) {
+        if (categoryArray[i].id === idToLookFor) {
+            return(categoryArray[i].colors);
+        }
+    }
+}
```
HenriChabert: Simplification
```suggestion
function findId(data, idToLookFor) {
    return data.palettes.find(p => p.id === idToLookFor).colors;
}
```

resource/gradient-color-scale.js
```diff
@@ -0,0 +1,58 @@
...
+    let sample = findId(dkuColorPaletteContinuous, colorPalette);
+    let myColor = d3.scaleQuantize()
+        .range(sample)
+        .domain([0, 10]);
+    gradient = {};
+    for (let i=0; i <= 9; i++) {
+        gradient[i/10] = myColor(i);
+    }
+    return gradient
+}
```
HenriChabert: DRY: Use that kind of function
```javascript
function paletteToGradient(colorRange, domain) {
    let colorScale = d3.scaleQuantize()
        .range(colorRange)
        .domain(domain);
    gradient = {};
    for (let i=0; i <= 9; i++) {
        gradient[i/10] = colorScale(i);
    }
    return gradient
}

function convertBuiltInPaletteToGradient(colorPalette){
    return paletteToGradient(findId(dkuColorPaletteContinuous, colorPalette), [0, 10]);
}
```

resource/gradient-color-scale.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Convert the name of the palette to the leaflet heat map gradient object
+ * @param colorPalette: Name of the input colorPalette
+ * @returns {{}}
+ */
+function convertPaletteToGradient(colorPalette){
+    var isFirstLetterUppercase = /^[A-Z]/.test(colorPalette);
+    if (isFirstLetterUppercase) return convertColorBrewerPaletteToGradient(colorPalette)
+    else return convertBuiltInPaletteToGradient(colorPalette)
+}
```
HenriChabert: Great ternary operator for this kind of things:
```suggestion
   return isFirstLetterUppercase ? convertColorBrewerPaletteToGradient(colorPalette) : convertBuiltInPaletteToGradient(colorPalette)
}
```

Or to be even more concise
```suggestion
    return (isFirstLetterUppercase ? convertColorBrewerPaletteToGradient : convertBuiltInPaletteToGradient)(colorPalette);
}
```

resource/dku-helpers.js
```diff
@@ -0,0 +1,85 @@
...
+    if (webAppConfig['source'] === webAppConfig['target']) {
+        throw Error("Columns must be different")
+    }
+}
+
+function quadraticScalingFunction(min, max, total, value) {
+    if (max === min) return 0;
+    let scale = 1 / (max - min);
+    return Math.pow(Math.max(0,(value - min)*scale), 2);
+}
```
HenriChabert: Unused functions

resource/dku-helpers.js
```diff
@@ -0,0 +1,85 @@
...
+
+
+dataiku.webappBackend = (function() {
+    function getUrl(path) {
+        return dataiku.getWebAppBackendUrl(path);
+    }
+
+    function dkuDisplayError(error) {
+        console.error("Backend error: ", error);
+    }
```
HenriChabert: Unused function

resource/config-event.js
```diff
@@ -0,0 +1,190 @@
...
+
+    this.getConfigAsJson = function () {
+        return {
+            "datasetName": _datasetName,
+            "detailsColumnName": _detailsColumnName,
+            "filters": _filters,
+            "geopointColumnName": _geopointColumnName,
+            "tooltipColumnName": _tooltipColumnName
+        }
+    }
```
HenriChabert, line 209: Unused function

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+
+    Object.defineProperty(this, 'mapPointer', {
+        get: function(){ return _mapPointer;}
+    });
+
+    /**
+     * Part of the initialisation of the chart
+     */
+    this.initialiseMap = function(){
+        _mapPointer = L.map(_mapId).setView([42, -17], 1);
```
HenriChabert: Use a constant to define default map coords
```suggestion
        const defaultCoords = [42, -17];
        _mapPointer = L.map(_mapId).setView(defaultCoords, 1);
```

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+     * Part of the initialisation of the chart
+     */
+    this.addLeafletLayer = function(){
+        L.tileLayer('http://{s}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}@2x.png', {
+            maxZoom: 18,
+            attribution: 'Map data &copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, &copy; <a href="http://cartodb.com/attributions">CartoDB</a>',
+            id: 'light_all',
+            tileSize: 512,
+            zoomOffset: -1
+        }).addTo(_mapPointer).set;
```
HenriChabert: Put this in a function. If you change the maxZoom for example, you will have to change it in two different places which is not optimal.

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+     *      1 - Closest point from cursor (None if no point at all)
+     *      2 - The set of points in the rectangular search window with fixed search radius (None if no points in search window)
+     * @returns {*[]|(*|number|bigint)[]}
+     */
+    this.getNeighbors = function(lat, long, mode=1){
+        // If no quadtree defined or quadtree is empty return no neighbors
+        if (!_quadtree || _quadtree.size() === 0){
+            return [];
+        }
+        if (mode === 1){
```
HenriChabert: As discussed, use a different branch to handle a different mode so that the one going to prod is as clean as possible

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+                    .duration('50')
+                    .attr("fill", "red")
+                    .attr('r', 10)
+            })
+            .on('mouseout', function() { // handle
+                _tooltip.style("visibility", "hidden");
+                d3.select(this).transition()
+                    .duration('150')
+                    .attr("r", 5)
+            })
```
HenriChabert: While testing the webapp, it seems that sometimes, the tooltip is not hidden when leaving the circle. I did not really investigate why because it remains quite smooth but I don' t know if it is the expected behaviour

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+                .attr("cy", function(d){ return _mapPointer.latLngToLayerPoint([d.lat, d.long]).y })
+        })
+    };
+
+    /**
+     * Render the heatmap. Take into account pre-stored geospatial data and visual layout parameters.
+     */
+    this.render = function(){
+        _quadtree = d3.quadtree()
+            .extent([[-100, -100], [100, 100]])
```
HenriChabert: The min max range is `[[-90, -180], [90, 180]]`.

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+function formatTooltip(tooltip){
+    if (tooltip && Object.keys(tooltip).length === 0 && tooltip.constructor === Object) {
+        return ""
+    }
+    var formattedString = "<hr>";
+    for (var key of Object.keys(tooltip)) {
+        formattedString += key+": ";
+        formattedString += "<b>"+tooltip[key]+"</b>";
+        formattedString += "<br>";
+    }
```
HenriChabert: Better to use this syntax
```suggestion
    let formattedString = "<hr>";
    for (const [key, value]  of Object.entries(tooltip)) {
        formattedString += `${key}: <b>${value}</b><br>`
    }
```

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+        console.log("Call to displayLocal ...");
+        d3.selectAll("circle").remove();
+        // Define the d3 dynamics around the geospatial data points (rendering and events)
+        _svgPointer.selectAll("myCircles")
+            .attr("pointer-events", "visible")
+            .data(_closestMarker)
+            .enter()
+            .append("circle")
+            .attr("cx", function(d) {
+                _tooltip.html("<div>Lon: <strong>" + d.long + "</strong><br>Lat: <strong>"+d.lat+"</strong>" +
```
HenriChabert: To match the scatter plot behaviour, you should also display in the tooltip the `Details` information. Don' t you have the Details infos line 215 when  searching for the closest neighbour?

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+            if (!closestMarker) {
+                console.warn('Quadtree not able to find closest point');
+            }
+            _closestMarker = [];
+            console.log("Closest marker:", closestMarker);
+            closestMarker.forEach(function (item, index) {
+                _closestMarker.push({lat: item[0], long: item[1], tooltip: item[3]});
+            });
+            console.log("_closestMarker:", _closestMarker);
+            this.displayLocal(_tooltip);
```
HenriChabert: Unused parameter : _tooltip

resource/geo-density-chart.js
```diff
@@ -0,0 +1,379 @@
...
+        return ""
+    }
+    var formattedString = "<hr>";
+    for (var key of Object.keys(tooltip)) {
+        formattedString += key+": ";
+        formattedString += "<b>"+tooltip[key]+"</b>";
+        formattedString += "<br>";
+    }
+    return formattedString
+}
```
HenriChabert, line 364: As discussed, here are my changes to use the native popup from leaflet rather than the custom one. It can surely be improved with these [options](https://leafletjs.com/reference-1.7.1.html#popup-option).

```javascript
function GeoDensityChart(){
    // ....
    this.displayLocal = function(){

        console.log("Call to displayLocal ...");
        d3.selectAll("circle").remove();
        // Define the d3 dynamics around the geospatial data points (rendering and events)
        var popup = L.popup()
            .setLatLng(L.latLng(_closestMarker[0].lat, _closestMarker[0].long))
            .setContent(formatTooltip(_closestMarker[0]));
        popup.openOn(_mapPointer);
        _svgPointer.selectAll("myCircles")
            .attr("pointer-events", "visible")
            .data(_closestMarker)
            .enter()
            .append("circle")
            .attr("cx", function(d) {
                return _mapPointer.latLngToLayerPoint([d.lat, d.long]).x
            })
            .attr("cy", function(d) {
                return _mapPointer.latLngToLayerPoint([d.lat, d.long]).y
            })
            .attr("id", "circleBasicTooltip")
            .on('mouseover', function() { // handle the event user mouse is over the circle data point
                popup.openOn(_mapPointer);
                d3.select(this).transition()
                    .duration('50')
                    .attr("fill", "red")
                    .attr('r', 10)
            })
            .on('mouseout', function() { // handle
                d3.select(this).transition()
                    .duration('150')
                    .attr("r", 5)
            })
            // potentially useless
            .attr("r", 5)
            .attr("stroke", "black")
            .attr("stroke-width", 3)
            .attr("fill-opacity", .4)
            .attr("fill", "transparent")
    };
}
function formatTooltip(closestMarker){
    let tooltipHTML = `<div>Lon: <strong>${closestMarker.long}</strong><br>Lat: <strong>${closestMarker.lat}</strong>`
    if (!closestMarker?.tooltip || !Object.keys(closestMarker.tooltip).length) {
        return tooltipHTML
    }
    tooltipHTML += "<hr>"
    for (const [key, value]  of Object.entries(closestMarker.tooltip)) {
        tooltipHTML += `${key}: <b>${value}</b><br>`
    }
    return tooltipHTML
}
```

You can then remove all references to _tooltip and remove the div.
Furthermore, check what to do with _closestMarker, weather you should leave it as a lsit or as a dict because in this mode, it will always be unique

python-lib/utils/df_column_naming.py
```diff
@@ -0,0 +1,26 @@
+import logging
+
+from collections import defaultdict
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO, format='%(asctime)s Custom Chart Geospatial Density  | %(levelname)s - %(message)s')
+
+
+def get_new_column_name(df, prefix=""):
```
HenriChabert: I had some trouble understanding what this function was doing despite the docstring. I would rather use the following signature : `make_column_name_unique(df, column_name)`

python-lib/utils/df_column_naming.py
```diff
@@ -0,0 +1,26 @@
...
+    name_is_taken = defaultdict(bool)
+    columns = list(df.columns)
+    for column in columns:
+        name_is_taken[column] = True
+    if not name_is_taken[prefix]:
+        return prefix
+    for suffix in range(len(columns)+1):
+        generated_name = prefix + "_" + str(suffix)
+        if not name_is_taken[generated_name]:
+            return generated_name
```
HenriChabert, line 27: I don't really understand why you do all that, maybe I miss something!
I would have personally done the following
```suggestion
    columns = list(df.columns)
    if not prefix in columns:
        return prefix
    for suffix in range(len(columns)+1):
        generated_name = prefix + "_" + str(suffix)
        if not generated_name in columns:
            return generated_name
```

python-lib/dku_data_processing/filtering.py
```diff
@@ -0,0 +1,166 @@
+"""
```
HenriChabert, line 1: About the problem you told me about filtering, the error message when deselecting all fields of a Text were : `Backend error: <html> <head><title>414 Request-URI Too Large</title></head> <body> <center><h1>414 Request-URI Too Large</h1></center> <hr><center>nginx</center> </body> </html> . Check backend log for more information.`
This means that the URI you use to call the backend is too long, which is something local as URI max length is more or less 2000 characters. Actually, this comes from the fact that you are using a GET to call the URI, meaning all the parameters are passed through the querystring, and thus, when deselecting a long list of fields, it is a looooooooong URI that is called.
The fix for that is rather to use a POST, in which you will be able to pass as many data as you want in the body of the POST, and you will not get this error anymore.

resource/dku-helpers.js
```diff
@@ -0,0 +1,85 @@
...
+
+    function get(path, args={}, displayErrors=true) {
+        return fetch(`${getUrl(path)}?${$.param(args)}`, {
+            method: 'GET',
```
HenriChabert: You should use a method POST as the number of characters in an URI is limited to about 2,000. See the comment on filtering.py for more information

python-lib/dku_data_processing/format.py
```diff
@@ -0,0 +1,164 @@
...
+
+    if detail is not None:
+        columns_to_retrieve.add(detail)
+
+    if (filters is not None) and (filters != []):
+        for col_ in filters:
+            columns_to_retrieve.add(col_['column'])
+
+    # Sampling should be the same as DSS by default
+    df = dss_dataset.get_dataframe(sampling='head', limit=10000, columns=columns_to_retrieve)
```
HenriChabert: Fix for error white date filtering
```suggestion
    df = dss_dataset.get_dataframe(sampling='head', limit=10000)
    df = df[list(columns_to_retrieve)]
```
It seems that it is DSS Api that has a problem, this way will always work

python-lib/dku_data_processing/filtering.py
```diff
@@ -0,0 +1,166 @@
...
+    conditions = []
+    excluded_values = []
+    for k, v in filter_['excludedValues'].items():
+        if k != DKU_NO_VALUE:
+            if v:
+                excluded_values += [k]
+        else:
+            if v:
+                conditions += [~df[filter_['column']].isnull()]
+    if len(excluded_values) > 0:
```
HenriChabert: I think it is a good practice to rather use the syntax:
```suggestion
    if excluded_values:
```
Indeed, if ever excluded_values is Null, it will not generate an error but just not enter into the condition. What do you think?

python-lib/dku_data_processing/format.py
```diff
@@ -0,0 +1,164 @@
...
+            return float(x)
+        else:
+            return None
+
+    new_longitude_column = get_new_column_name(df, "longitude")
+    new_latitude_column = get_new_column_name(df, "latitude")
+
+    # Parse longitude and latitude from WKT geopoint column
+    parsed_lat_long_df = df[geopoint].apply(wkt_parser)
+    df[new_longitude_column] = parsed_lat_long_df.apply(lambda x: x[0]).apply(lambda x: convert(x))
```
HenriChabert: It is redoundant to use a lambda function
```suggestion
    df[new_longitude_column] = parsed_lat_long_df.apply(lambda x: x[0]).apply(convert)
```

python-lib/dku_data_processing/format.py
```diff
@@ -0,0 +1,164 @@
...
+    else:
+        df[new_detail_column] = normalize(df[detail].values)
+    if (detail is not None) and (df.dtypes[detail] not in ['int64', 'float64']):
+        raise ValueError("The detail column should be a int or float type.")
+
+    def convert(x):
+        if x is not None:
+            return float(x)
+        else:
+            return None
```
HenriChabert, line 92: Simplification proposal: 
```suggestion
    convert = lambda x: x and float(x)
```

HenriChabert (changes requested):
As usual, very good work!

Most of my comments are syntax based or just about writing more readable code.
I also fixed some errors we spoke about, everything is open to discussion and can be improved.

Do not hesitate if you have questions :)

js/leftBarParamsModule.js
tdesfont: Agreed, done in c48e478

plugin.json
tdesfont: Done in 18a9c5a

plugin.json
tdesfont: Done in 18a9c5a

python-lib/dku_data_processing/filtering.py
tdesfont: Very nice catch, done in 524316c, should investigate how POST request in dip webapp backend handle params but works after changing methods to POST

python-lib/dku_data_processing/filtering.py
tdesfont: Agreed and done in 4cf4276

python-lib/dku_data_processing/format.py
tdesfont: Done in 4cf4276

python-lib/dku_data_processing/filtering.py
tdesfont: **Should change the way data is sent through the request before merge**

python-lib/dku_data_processing/format.py
tdesfont: Done in e0e3559

python-lib/dku_data_processing/format.py
tdesfont: Done in bdaadc9

python-lib/utils/df_column_naming.py
tdesfont: So the use of the defaultdict(bool) is for optimization, it is expected to give a linear instead of a quadratic computation time wrt the number of columns, but it might be over engineering

python-lib/utils/df_column_naming.py
tdesfont: Changed in 62fa6a6

resource/config-event.js
tdesfont: The function getConfigAsJson is marked as unused in my IDE but is actually called in 
`function updateCoreData(configEvent, chartHandler) {

    console.log("Starting query of server to fetch core data ....");
    dataiku.webappBackend.get('get_geo_data', {
        "config": JSON.stringify(configEvent.getConfigAsJson())
    })
        .then(function(data){`

resource/dku-helpers.js
tdesfont: Removed unused functions in 3511fcb

resource/dku-helpers.js
tdesfont: Done in 1ee5226

resource/geo-density-chart.js
tdesfont: Done in 3e4b842

resource/geo-density-chart.js
tdesfont: Removed in 8bb75f9

resource/geo-density-chart.js
tdesfont: Indeed the tooltip is sometimes stuck and the link with the mouse is broken, don't know why and as you said, when the mouse event is detected again the tooltip follows showing expected behaviour. I will try to understand why but it might be difficult to reproduce

resource/geo-density-chart.js
tdesfont: This method was subject to question, it seems that we can get rid of this call, removed in ba0a073

resource/geo-density-chart.js
tdesfont: Done in c71463d

resource/geo-density-chart.js
tdesfont: Very neat, thanks a lot ! Done in a17d92f

resource/geo-density-chart.js
tdesfont: **After introduction of the leaflet popup, the mouse outside the map leads to popup of the closest marker even if mouse is not over it, will need to investigate a bit**

resource/leftBarParamsTemplate.html
tdesfont: You are a JavaScript maestro ! Done in 2af2422

resource/leftBarParamsTemplate.html
tdesfont: Done

webapps/density-custom-chart/app.js
tdesfont: Called removed in 9a9805b

webapps/density-custom-chart/app.js
tdesfont: Changed in cdc5a27

webapps/density-custom-chart/app.js
tdesfont: Done in 4d1d2f0

webapps/density-custom-chart/app.js
tdesfont: Included in e0b0d9e

webapps/density-custom-chart/app.js
tdesfont: Done in adbc345

webapps/density-custom-chart/app.js
tdesfont: Done in adbc345

webapps/density-custom-chart/app.js
tdesfont: Done in 393a5b8, need to check variables scope

webapps/density-custom-chart/app.js
tdesfont: Done in 6e23f50

webapps/density-custom-chart/app.js
tdesfont: Done in a7e80c6

webapps/density-custom-chart/body.html
tdesfont: Done in bc1a9ee

webapps/density-custom-chart/body.html
tdesfont: Done in 64a3e95

webapps/density-custom-chart/body.html
tdesfont: Done in 64a3e95

webapps/density-custom-chart/style.css
tdesfont: This style is actually not used, removed in 530f272

webapps/density-custom-chart/style.css
tdesfont: This one seems to be blue, skipping it

tdesfont: Where are those vars ?

webapps/density-custom-chart/style.css
tdesfont: Removed in 2d928cf

webapps/density-custom-chart/style.css
tdesfont: Removed

webapps/density-custom-chart/style.css
tdesfont: Removed tooltip style unless needed, it seems that introduction of leaflet tooltip make this obsolete

webapps/density-custom-chart/webapp.json
tdesfont: Well Known Text, according to previous development  we should say WKT since its pretty common, will keep this in mind if global change occurred on that convention

webapps/density-custom-chart/webapp.json
tdesfont: Indeed, this point was also noticed during product oriented sessions but we decided to leave it as is since we plan to make this chart part of the native ones

python-lib/dku_data_processing/filtering.py
tdesfont: Done in b8c03ae

resource/geo-density-chart.js
tdesfont: Done in 64ced85

tdesfont: Might be enhanced through default/optional parameters of the function generateDefaultConfig instead of setting them manually in the json

tdesfont (commented):
Response to the comments

webapps/density-custom-chart/style.css
tdesfont: Removed

webapps/density-custom-chart/style.css
tdesfont: Removed

webapps/density-custom-chart/style.css
tdesfont: Removed

webapps/density-custom-chart/app.js
tdesfont: Done

resource/geo-density-chart.js
tdesfont: Fixed

resource/geo-density-chart.js
tdesfont: Thanks for the precision though

resource/geo-density-chart.js
tdesfont: I added the detail but it is a bit overkill for now (for loop on the records to add the same information), I need to refactor this, any suggestion welcome

resource/dku-helpers.js
tdesfont: Changed the request to post with args in body

resource/gradient-color-scale.js
tdesfont: Done in 0bed5c7

resource/gradient-color-scale.js
tdesfont: Done in 0bed5c7

resource/gradient-color-scale.js
tdesfont: Done in 0bed5c7

webapps/density-custom-chart/style.css
HenriChabert: You can find them in the following file: http://localhost:11200/static/public/styles/1.0.0/variables.css

resource/geo-density-chart.js
HenriChabert: Yop, better indeed

PR #8, opened 2021-02-10
[ch60792] [density chart] Fix bug when publishing dashboard

tdesfont:
[ch60792](https://app.clubhouse.io/dataiku/story/60792/density-chart-backend-error-when-you-publish-a-density-chart-with-a-filter-to-a-dashboard)

filtering.py: Set default value to left/right bounds for filters, changed the empty dataframe to a log instead of an error
format.py: Refactored code to allow test on dataframe processing, handle case with normalizer on one single value
 test_geodata.py: Refactored the unit tests to test only the dataframe processing part which is sufficient in our case
 test_dataset_retrieval.py: Deleted unused file


tdesfont:
Closing this pull request to open a cleaner one

PR #9, opened 2021-02-10
[ch60793] [density chart] Set higher default quality of the tile maps

tdesfont:
[ch60793](https://app.clubhouse.io/dataiku/story/60793/density-chart-background-map-is-blurry)


python-lib/dku_data_processing/filtering.py
```diff
@@ -1,12 +1,19 @@
...
 
 DKU_NO_VALUE = '___dku_no_value___'
 
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO, format='%(asctime)s Custom Chart Geospatial Density  | %(levelname)s - %(message)s')
+
 
 def numerical_filter(df, filter_):
     conditions = []
+    if "minValue" not in filter_:
```
tdesfont: Related to dashboards, make sure not to pass None as range limit

python-lib/dku_data_processing/format.py
```diff
@@ -17,7 +13,12 @@ def normalize(x):
     """
     Will normalize data to have 75% of the values between 0 and 1
     """
-    x_scaled = (x - np.percentile(x, 12.5)) / (np.percentile(x, 87.5) - np.percentile(x, 12.5))
+    perc_left = np.percentile(x, 12.5)
```
tdesfont: Handle the case where the serie x length is one.

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
         return result[0]
 
 
+def extract_df(df, detail, filters, geopoint, tooltips):
```
tdesfont, line 35: Refactoring of the backend function call to geocolumn

python-lib/dku_data_processing/format.py
```diff
@@ -71,42 +126,8 @@ def fetch_geo_data(dss_dataset: Dataset, geopoint, detail, tooltips, filters):
...
-
-    if (detail is not None) and (df.dtypes[detail] not in ['int64', 'float64']):
-        raise ValueError("The detail column should be a int or float type")
-
-    def convert(x):
-        if x is not None:
-            return float(x)
-        else:
-            return None
+    geodata_object = extract_df(df, detail, filters, geopoint, tooltips)
```
tdesfont, line 137: Extract backend function to process dataframe instead of DSS dataset (allow python unit tests)

resource/geo-density-chart.js
```diff
@@ -71,7 +71,7 @@ function GeoDensityChart(){
     };
 
     this.addLeafletLayer = function(){
-        L.tileLayer('http://{s}.basemaps.cartocdn.com/dark_all/{z}/{x}/{y}.png', {
+        L.tileLayer('http://{s}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}@2x.png', {
```
tdesfont: Change tile service url adding @2x to increase background image precision

python-lib/dku_data_processing/filtering.py
```diff
@@ -1,12 +1,19 @@
...
-
+import logging
 
 DKU_NO_VALUE = '___dku_no_value___'
 
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO, format='%(asctime)s Custom Chart Geospatial Density  | %(levelname)s - %(message)s')
+
 
 def numerical_filter(df, filter_):
```
Marsobad, line 11: you could use a `filter_.get("minValue", -float('inf'))` to simplify the code

python-lib/dku_data_processing/filtering.py
Marsobad, line 13: you could use a `filter_.get("minValue", -float('inf'))` to simplify the code

python-lib/dku_data_processing/format.py
```diff
@@ -17,7 +13,12 @@ def normalize(x):
     """
     Will normalize data to have 75% of the values between 0 and 1
     """
-    x_scaled = (x - np.percentile(x, 12.5)) / (np.percentile(x, 87.5) - np.percentile(x, 12.5))
+    perc_left = np.percentile(x, 12.5)
+    perc_right = np.percentile(x, 87.5)
```
Marsobad, line 17: `right_percentile` and `left_percentile` would sound more meaningful to me rather than `perc_left` and `perc_right`

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
...
+
+    def convert(x):
+        if x is not None:
+            return float(x)
+        else:
+            return None
+
+    # Parse longitude and latitude from WKT geopoint column
+    parsed_lat_long_df = df[geopoint].apply(wkt_parser)
+    df['longitude_'] = parsed_lat_long_df.apply(lambda x: x[0]).apply(lambda x: convert(x))
```
Marsobad, line 69: Be careful with collisions : what will you do if there's already a "longitude_" column?

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
...
+    df['latitude_'] = parsed_lat_long_df.apply(lambda x: x[1]).apply(lambda x: convert(x))
+    df = df.drop(geopoint, axis=1)
```
Marsobad, line 71: Is this line really necessary since you apply another filter at line 81: 
`df = df[['latitude_', 'longitude_', 'detail_']]`

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
...
+    # Handle normalizing of detail column
+    if (detail is None) or (df.dtypes[detail] not in ['int64', 'float64']):
+        df['detail_'] = 1
+    else:
+        df['detail_'] = normalize(df[detail].values)
+    if (detail is not None) and (df.dtypes[detail] not in ['int64', 'float64']):
+        raise ValueError("The detail column should be a int or float type")
+
+    def convert(x):
+        if x is not None:
```
Marsobad, line 62: Nitpicking - you can replace 'if x is not None` with `if x`(but maybe if x is not none is more readable, up to you)

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
...
+        df = filter_dataframe(df, filters)
+    if df.empty:
+        return []
+
+    # Handle normalizing of detail column
+    if (detail is None) or (df.dtypes[detail] not in ['int64', 'float64']):
+        df['detail_'] = 1
+    else:
+        df['detail_'] = normalize(df[detail].values)
+    if (detail is not None) and (df.dtypes[detail] not in ['int64', 'float64']):
```
Marsobad, line 58: replace `if detail is not None` with `if detail`

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
...
+
+    # Parse longitude and latitude from WKT geopoint column
+    parsed_lat_long_df = df[geopoint].apply(wkt_parser)
+    df['longitude_'] = parsed_lat_long_df.apply(lambda x: x[0]).apply(lambda x: convert(x))
+    df['latitude_'] = parsed_lat_long_df.apply(lambda x: x[1]).apply(lambda x: convert(x))
+    df = df.drop(geopoint, axis=1)
+
+    # Handle tooltip
+    tooltip_column_to_keep = set()
+    if (tooltips is not None) and (tooltips != []):
```
Marsobad: you can replace the line 75 with `if tooltips`

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
...
+    df['tooltip_'] = tooltip_df.to_dict(orient='records')
+    df = df.rename(columns={"latitude_": "lat", "longitude_": "long", "detail_": "detail", "tooltip_": "tooltip"})
+    df = df.dropna(subset=['lat', 'long'])
+
+    geodata_object = df.to_dict(orient='records')
+
+    return geodata_object
+
+
 def fetch_geo_data(dss_dataset: Dataset, geopoint, detail, tooltips, filters):
```
Marsobad: inconsistent typing : only dss_dataset is typed here and the Dataset object raises import issues in the unit tests
You could remove the type `Dataset` in the final version

python-lib/dku_data_processing/format.py
Marsobad: Is this line really necessary since you apply another filter at line 81?: 
`df = df[['latitude_', 'longitude_', 'detail_']]`

python-lib/dku_data_processing/format.py
Marsobad: Be careful with collisions : what will you do if there's already a "longitude_" column?

python-lib/dku_data_processing/format.py
Marsobad: To me, `right_percentile` and `left_percentile` would sound more meaningful than `perc_left` and `perc_right`

python-lib/dku_data_processing/filtering.py
Marsobad: you could use a `filter_.get("minValue", -float('inf'))` to simplify the code

python-lib/dku_data_processing/format.py
```diff
@@ -31,6 +32,62 @@ def wkt_parser(wkt_point):
...
+    Extract and select only necessary data to send to front-end.
+    Handle the filtering, normalizing, parsing and the tooltip.
+
+    :param detail:
+    :param df:
+    :param filters:
+    :param geopoint:
+    :param tooltips:
+    """
+    # Handle filtering
```
Marsobad: I feel that you could turn each commented block into a function doing what your comment says

tests/python/unit/test_geodata.py
```diff
@@ -12,23 +13,14 @@ def convert_numpy_int64_to_int(o):
...
-def test_extract_lat_long_invalid():
-    coordinates = [None, 'POINT(-73.98377 40.75362)', 'POINT(-73.9419 40.80902)']
-    lat, long_ = extract_lat_long(coordinates)
-    assert False
+    assert result[0] == '-45'
+    assert result[1] == '59'
 
 
-def test_format_geodata():
+def test_filtering_lead_to_empty_df():
```
Marsobad, line 20: This test returns an error because of the `dataiku`package in `format.py`

python-lib/dku_data_processing/format.py
```diff
@@ -1,7 +1,3 @@
-import dataiku
-import dataikuapi
-import pdb
-
 from dataiku import Dataset
```
Marsobad: remove dataiku import

Marsobad (changes requested):
Perfect fix for the CH cart ! The quality issue is fixed
Some extra remarks:
- Don't hesitate to replace `if sth is not None` with `if sth`
- In requirements for the Python tests, add the packages required by your unit tests including numpy and pandas
- For unit tests, I think you should add other filtering tests with filters containing an `ALPHANUM_FACET`. I noticed a bug that I described in [[ch61148](https://app.clubhouse.io/dataiku/story/61148/density-chart-error-when-a-numerical-filter-is-treated-as-text)]. It would also be nice to test for invalid input parameters as you did previously with `test_format_geodata_invalid_geocolumn`

python-lib/dku_data_processing/filtering.py
tdesfont: I used the dict get but changed the logic to enhance the code in 45e351b

python-lib/dku_data_processing/format.py
tdesfont: Done in 261e6bf

python-lib/dku_data_processing/format.py
tdesfont: Done in 7420b8c

python-lib/dku_data_processing/format.py
tdesfont: Done

python-lib/dku_data_processing/format.py
tdesfont: Removed Dataset typing in 261e6bf

python-lib/dku_data_processing/format.py
tdesfont: Removed drop geopoint in 683be05

python-lib/dku_data_processing/format.py
tdesfont: Implemented an intermediary column naming to solve the collisions issue

tests/python/unit/test_geodata.py
tdesfont: Deleted the Dataset typing leading to the error

python-lib/dku_data_processing/format.py
tdesfont: Will leave this out of this PR but duly noted

PR #10, opened 2021-02-10
[ch60779] [density chart] minor: Avoid blinking map when no actual difference is sent 

tdesfont:
[ch60779](https://app.clubhouse.io/dataiku/story/60779/density-chart-chart-is-reloaded-when-showing-hiding-parameter-sections)
Avoid blinking map by comparing previous and next maptile ids, also setting maptile quality to highest quality

### DIFF
--- a/resource/geo-density-chart.js
+++ b/resource/geo-density-chart.js
@@ -86,25 +86,26 @@ function GeoDensityChart(){
             if (layer.options.id){
                 console.log("[MapTile] Received previous maptile id:", layer.options.id);
                 let url;
-                let idTileMap;
+                let newBaseMap;
                 console.log("[MapTile] Received asked maptile id", mapTile);
                 switch (mapTile) {
                     case 'cartodb-positron':
                         url = 'http://{s}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}@2x.png';
-                        idTileMap = "light_all";
+                        newBaseMap = "light_all";
                         break;
                     case 'cartodb-dark':
                         url = 'http://{s}.basemaps.cartocdn.com/dark_all/{z}/{x}/{y}@2x.png';
-                        idTileMap = "dark_all";
+                        newBaseMap = "dark_all";
                         break;
                 }
-                if (layer.options.id !== idTileMap){
-                    console.log("[MapTile] Building new map tilelayer with id", idTileMap);
+                let previousBaseMap = layer.options.id;
+                if (previousBaseMap !== newBaseMap){
+                    console.log("[MapTile] Building new map tilelayer with id", newBaseMap);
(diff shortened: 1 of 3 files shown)


resource/geo-density-chart.js
```diff
@@ -84,23 +84,31 @@ function GeoDensityChart(){
         // iterate in the layers of the map
         _mapPointer.eachLayer(function(layer){
             if (layer.options.id){
```
tdesfont, line 86: Select the map layer using the key layer.options.id which is only available for map

resource/geo-density-chart.js
```diff
@@ -84,23 +84,31 @@ function GeoDensityChart(){
...
+                console.log("[MapTile] Received previous maptile id:", layer.options.id);
                 let url;
-                _mapPointer.removeLayer(layer);
+                let idTileMap;
```
Marsobad: You could perhaps rename `idTileMap` to sth like `newMapBackground` because for me, this name makes it difficult to understand that it's actually the map background colour.

resource/geo-density-chart.js
```diff
@@ -84,23 +84,31 @@ function GeoDensityChart(){
...
                         break;
                 }
-                L.tileLayer(url, {
-                    maxZoom: 18,
-                    attribution: 'Map data &copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, &copy; <a href="http://cartodb.com/attributions">CartoDB</a>',
-                    id: 'light_all',
-                    tileSize: 512,
-                    zoomOffset: -1
-                }).addTo(_mapPointer).set;
+                if (layer.options.id !== idTileMap){
```
Marsobad: To make it easier to understand, I suggest to name the boolean `layer.options.id !== idTileMap`
Perhaps : `isBackgroundUpdated` or `hasBackgroundChanged`or `backgroundChange`

Marsobad (changes requested):
- This fix solves the issue when you change parameters 
- I noticed that the view is re-scaled when you modify a **filter** or a **tooltip** while it's not the case for the scatter map. Is it intentional?

resource/geo-density-chart.js
tdesfont: I renamed `idTileMap` to `newBaseMap`, I think the light_all is more than just the color of the map it can be an id for an exotic map background

tdesfont: Done in d2c7118

resource/geo-density-chart.js
tdesfont: Done in d2c7118, add precision on the two compared variable

Marsobad (approved):
Looks great to me !

PR #11, opened 2021-02-11
[ch60785] [density chart] Stick to the leaflet popup UX in Scatter Plot

tdesfont:
[ch60785](https://app.clubhouse.io/dataiku/story/60785/density-chart-enhance-ux-of-tooltip-like-in-scatter-chart)

app.js: Replicate the same style properties for the tooltip
style.css: Removed unused style class
geo-density-chart.js: Replicate same style

### DIFF
--- a/resource/geo-density-chart.js
+++ b/resource/geo-density-chart.js
@@ -285,11 +285,11 @@ function formatTooltip(tooltip){
     if (!tooltip){
         return
     }
-    var formatedString = "<hr>";
+    var formattedString = "<hr>";
     for (var key of Object.keys(tooltip)) {
-        formatedString += key+": ";
-        formatedString += "<b>"+tooltip[key]+"</b>";
-        formatedString += "<br>";
+        formattedString += key+": ";
+        formattedString += "<b>"+tooltip[key]+"</b>";
+        formattedString += "<br>";
     }
-    return formatedString
+    return formattedString
 }
--- a/resource/geo-density-chart.js
+++ b/resource/geo-density-chart.js
@@ -282,7 +282,7 @@ function GeoDensityChart(){
  * @returns {string}
  */
 function formatTooltip(tooltip){
-    if (jQuery.isEmptyObject(tooltip)) {
+    if (tooltip && Object.keys(tooltip).length === 0 && tooltip.constructor === Object) {
         return ""
     }
     var formattedString = "<hr>";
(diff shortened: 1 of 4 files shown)


resource/geo-density-chart.js
```diff
@@ -285,13 +285,11 @@ function formatTooltip(tooltip){
     if (!tooltip){
         return
     }
-    var formatedString = "";
-    formatedString += "<div>";
+    var formatedString = "<hr>";
```
Marsobad: typo : the correct spelling is `formatted`

Marsobad (changes requested):
The tooltips look neater with this update !

I noted some inconsistencies with the scatter map :
- when only lon and lat are displayed, you should hide the bottom line 
- In the scatter map, tooltips are speech bubbles. In the density charts, tooltips are rectangles.
- lncrease the width of the bubbles?

resource/geo-density-chart.js
tdesfont: Done in 076d73e

tdesfont:
- [x] When only lon and lat are displayed, you should hide the bottom line: Done in c465eaa
- [ ] In the scatter map, tooltips are speech bubbles whereas in the density charts, tooltips are rectangles: Will add it as a card and don't plan to do it here. 
- [x] lncrease the width of the bubbles: Done in ae98ca7

resource/geo-density-chart.js
```diff
@@ -279,19 +279,17 @@ function GeoDensityChart(){
  * Format the tooltip to be displayed in HTML
  * tooltip = {"reviews_per_month": 0.92, "room_type": "Private Room"};
  * @param tooltip
- * @returns {number}
+ * @returns {string}
  */
 function formatTooltip(tooltip){
-    if (!tooltip){
-        return
+    if (jQuery.isEmptyObject(tooltip)) {
```
Marsobad: Are you sure about using jQuery? Granted, this solution is cleaner than[ native js code](https://stackoverflow.com/questions/679915/how-do-i-test-for-an-empty-javascript-object) but jQuery has become quite slow nowadays and I heard that most frontenders advised against it

Marsobad (commented):
Leaflet look good to me ! Beware of jQuery though, especially if you're not using it elsewhere

resource/geo-density-chart.js
tdesfont: I did not know that, thanks for pointing it out, I used an alternative in c5ef65a

Marsobad (approved):
All right ! I have nothing to add on my side

PR #12, opened 2021-02-16
[ch61233][density chart] Set sampling to the same configuration as in DSS

tdesfont:
[[ch61233]
](https://app.clubhouse.io/dataiku/story/61233/densitychart-set-sampling-to-head-10000)
Since we cannot get the sampling from DSS, we will give the default DSS configuration by default.

### DIFF
--- a/python-lib/dku_data_processing/format.py
+++ b/python-lib/dku_data_processing/format.py
@@ -69,7 +69,8 @@ def fetch_geo_data(dss_dataset: Dataset, geopoint, detail, tooltips, filters):
         for col_ in filters:
             column_to_retrieves.add(col_['column'])
 
-    df = dss_dataset.get_dataframe(limit=1000, columns=column_to_retrieves)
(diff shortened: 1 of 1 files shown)


python-lib/dku_data_processing/format.py
```diff
@@ -69,7 +69,8 @@ def fetch_geo_data(dss_dataset: Dataset, geopoint, detail, tooltips, filters):
         for col_ in filters:
             column_to_retrieves.add(col_['column'])
```
Marsobad: typo in `column_to_retrieves`==> `columns_to_retrieve`

Marsobad (commented):
LGTM ! - I just noted a typo

python-lib/dku_data_processing/format.py
tdesfont: Fixed in e2ae4af

